PR #1, opened 2015-07-28
Clarify criteria

domenic:
The main item here is to clarify that these are entrance criteria, not exit criteria.

If someone believes otherwise, it would be great if they could pull up minutes from previous meetings to support this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,6 +90,7 @@
       <ul>
         <li>Above
         <li>Initial spec text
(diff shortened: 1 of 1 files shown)


rwaldron:
Great work!

domenic:
Follow-up commit b3dae8e729d5a992e112f313fb8be38c6e969aea

PR #6, opened 2016-06-09
Clarify seemingly redundant criterium

mathiasbynens:
“The ECMAScript editor has signed off on the current spec text” is already a requirement for stage 3 and is thus covered by the “Above” criterium in stage 4.


allenwb:
I don't think this is redundant.  The intent is that the editor must sign off on the spec text that is being approved for stage 4.  This probably is not the same as the stage 3 text.

Perhaps this could be clarified by saying: "The editor has singed off on the final spec text"

mathiasbynens:
Patch updated.

chicoxyzzy:
Seems like this was fixed by https://github.com/tc39/process-document/commit/e0009705959360f35cf44468c5568e9d5bdbae0d

mathiasbynens:
Yep, it was.

PR #13, opened 2017-10-31
Add Stage 3.5

littledan:
This proposal attempts to solve a few problems with the current stage model:
- As a committee, we don't agree on what stage advancement means,
  particularly for reaching Stage 4. In the September 2017 TC39 meeting,
  one proposal reached Stage 4 for having one implementation behind a
  flag and another shipping in a "nightly" version of a browser, while
  another was blocked from Stage 4 even though it was shipping in two
  nightly versions, .
- Some implementations, both browsers and infrastructure, prefer to not
  release unflagged/without plugins until a proposal reaches Stage 4, to
  avoid dealing with instability and uncertainty in Stage 3 features.
- The gap between Stage 3 and Stage 4 is pretty large; many
  implementations wait until Stage 3 to review a proposal in detail,
  since from their perspective, prior to Stage 3, things are very
  volatile. However, we don't come to a particular point where we
  recognize that their feedback has been taken into account.

To address these issues, this proposal adds a "Stage 3.5". This stage
is intended to be reached once a significant amount of feedback from
implementation experience has been taken into account. The stage indicates
a greater degree of certainty, without requiring multiple shipping
implementations. With the addition of this intermediate stage, it's hoped
that stability can be communicated while taking off pressure to land in
the main specification, allowing the more conservative interpretations
of Stage 4 to not take away a signal from the committee for relative
stability of the proposal.

This stage isn't meant to be an extra barrier for anybody, but rather
a tool for champions to use when they feel appropriate. Stage 3.5
requirements are a subset of Stage 4 requirements, and it would still
be fine to go straight from 3 to 4, but 3.5 gives some intermediate
recognition of status.

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,13 +115,29 @@
     <td>Spec compliant
   </tr>
   <tr>
+    <td>3.5
+    <td>Advanced candidate
+    <td>Significant refinement has come from some implementation experience
+    <td>
+      <ul>
+        <li>Above
+        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, and merged
+        <li>An in-depth implementation has been prototyped which pass most acceptance tests
+        <li>Implementation(s) in build tools or behind a flag has led to some realistic programmer feedback.
+      </ul>
+    </td>
+    <td>The proposal is significantly solidified based on some usage and feedback
+    <td>Semi-final: Most changes as a result of implementation experience are integrated
+    <td>Limited: only those deemed critical based on implementation experience
+    <td>Spec compliant
+  </tr>
+  <tr>
     <td>4
     <td>Finished
     <td>Indicate that the addition is ready for inclusion in the formal ECMAScript standard
     <td>
       <ul>
         <li>Above
-        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, and merged
         <li>Two compatible implementations which pass the acceptance tests
         <li>Significant in-the-field experience with shipping implementations, such as that provided by two independent VMs
         <li>A pull request has been sent to <a href="https://github.com/tc39/ecma262">tc39/ecma262</a> with the integrated spec text


jamiebuilds:
Why not stage 4 and 5? To avoid invalidating everything already on the web?

ljharb:
I think this only has value if the "some implementations" - everyone for whom stage 4 is a requirement - agree to implement at 3.5. This includes all browsers and node, but imo also includes `eslint` (specifically, esprima/acorn).

Will all 4 browsers, at the least, commit to begin implementing proposals at 3.5, and no longer wait for 4?

msaboff:
I don't think this proposal addresses the underlying problem.  That is there isn't broad consensus as to what constitutes an "implementation".

As an implementor, I cannot see implementing and shipping a proposal in a fully released version of our JS engine / browser unless that proposal has fully advanced to the current stage 4.  We have no problem implementing something that goes in a nightly build or preview and feel that is a great time to iron out smaller details.  This happened recently with the named capture groups proposal wherein a small issue was raised and resolved among the current implementations.  Given there are now two implementations of named capture groups proposal, we can advance it to stage 4.

My concern is that this stage 3.5 would delay when implementors actually implement a proposal.  I still can't see shipping a stage 3.5 proposal in a shipping product and would want to wait to full stage 4 is achieved.  The committee would still want implementor's feed back after two or more implementations exist.  How does this proposal help?

To summarize my point of view, stage 3 is when multiple implementors implement a proposal.  For most, that is in some preview form.  When there are at least two implementations, and final review can be done, we can move that proposal to stage 4.  At that point (most) implementations ship.  I don't see how we can require most or all JS engines to implement at stage 3.5.  We don't have all browsers implementing ES2016, let alone ES2017.

littledan:
@msaboff 

Here's my suggestion for the workflow:

- By Stage 3, get a transpiler/polyfill implementation and some significant user feedback; take this all into account before moving to that stage.
- After Stage 3, browsers see that the feature is stable and begin implementing. They should feel comfortable shipping in "preview" releases/behind a flag at this point, but not "stable" releases.
- By the time two browsers have gotten some significant implementation experience and their feedback has been taken into account, Stage 3.5 is reached. Now, browsers should feel comfortable shipping in stable releases.
- By the time the feature is shipping in multiple browsers in stable for a bit, and any further kinks are worked out (e.g., maybe it turned out to have web compatibility issues that were only discovered when shipped to tons of users), Stage 4 is reached and the proposal can go into the standard.

Some further tweaks to encode this workflow:

- Let's go back to two native preview implementations for Stage 3.5, as you suggest.
- Require a transpiler/polyfill implementation to get to Stage 3, if it's feasible to do so.
- For Stage 3.5 and Stage 4, we ask for native browser implementations, not just any vague implementation.
- For Stage 4, we require that "shipping" means "shipping in browsers in stable" (not preview). If I understand correctly, this is @domenic's interpretation of what the process doc really says (so it wouldn't require textual changes).
- Maybe we actually move the test262 requirement to Stage 3.

If it doesn't seem to implementers like this would improve the workflow, I definitely do not want to pursue this.

I guess for V8, the equivalent of the preview release would be when V8 puts the feature in "staging", i.e. enabled by turning on experimental JS features. By the time something moves to Canary, it's already "shipping", on the train to get into everyone's stable release in a matter of months unless someone stops it.

Specifically enfrancising browsers here, and differentiating them from transpilers, has the important benefit of making sure we don't add things to the specification that will never ship in the browser consensus. Other web standards bodies get into ugly modes where they sometimes standardize things which never actually ship, or only ship in one browser; it's great that we've been able to mostly avoid that.

Today, if all the native implementations left the committee and we had polyfills, TypeScript and Babel present, I'm afraid we'd keep standardizing things that would only be implemented in those environments. If the browsers actually did stop implementing things, I think the committee's job would be to solve that problem by working with the browsers directly, rather than standardizing things into the air and hoping that this puts pressure on browsers to pick it up.

While there are some native implementations besides browsers (Node.js is not really one as it currently typically uses V8 or ChakraCore, which are maintained by browser vendors), browsers remain a very important environment for TC39. Other implementation experience is valuable, but it would be a very bad situation if we got out of sync with browsers.

@ljharb

I completely agree that it'd be great to encourage tools to implement at Stage 3.5 rather than waiting for Stage 4; this is part of the motivation.

@thejameskyle

I don't have a strong opinion here; this was out of conservatism. Maybe rebranding to Stage 4 and 5 would help towards what @ljharb was suggesting.

littledan:
Or, an alternative (much simpler) would be to clarify a couple things in the current stage model: Stage 4 requires native implementations, and it's referring to a preview release. Then, we're putting things in the main spec earlier than @domenic 's interpretation of the stage model.

msaboff:
> By the time two browsers have gotten some significant implementation experience and their feedback has been taken into account, Stage 3.5 is reached. Now, browsers should feel comfortable shipping in stable releases.

This all depends what you mean by "stable".  If this is a shipping release, e.g. Safari 11.0, then that would require the proposal to be stage 4 (in the standard).  If "stable" means preview, e.g. Safari Tech Preview 41, then stage 3 is fine.  Stage 3.5 doesn't add anything in my mind.

> Or, an alternative (much simpler) would be to clarify a couple things in the current stage model: Stage 4 requires native implementations, and it's referring to a preview release. Then, we're putting things in the main spec earlier than @domenic 's interpretation of the stage model.

This is the resolution I would prefer.  Again I cite what happened in stage 3 for Named Capture Groups, V8 implemented behind a flag followed later by JSC, delivered in a Safari Tech Preview.  An issue was identified and resolved.  Both of those implementations conform to the proposal after the change.  Stage 3 is complete and we're ready for stage 4.

Instead of complicating the staging process, lets clarify the gating criteria so that it can be uniformly applied.  Sounds like this might resurrect past discussion.  Maybe a review of those discussions is sufficient or a new discussion may be required.

domenic:
I haven't had time to read all the comments here, but our position is that we must not move things to stage 4 until something is shipping unflagged in two browsers intended for consumers (not preview releases). This necessitates two browsers shipping during stage 3.

Maybe we need a stage 3.5 for some people who have so far misinterpreted stage 3. I tend to think not, and instead we should clear up any confusion about stage 3. But we are not willing to put things in the spec (stage 4) until they are shipping successfully in two browsers.

We've compromised on this in the past, as a matter of political convenience or simply the more vocal Google representatives not being in the room at the right time. But if advancing stages requires consensus, this is the criteria we intend to apply, based on how we read the process document (and how we helped shape it).

littledan:
OK, it's clear this proposal doesn't solve the problem for anyone, so closing it.

PR #14, opened 2017-11-01
Strengthen Stage 3 and Stage 4 requirements

littledan:
- Require test262 tests out for review for Stage 3
- Require a polyfill/transpiler implementation for Stage 3
- Require native browser preview release, rather than "shipping experience"

This is an alternative to #13 and attempts to encode something more
closely related to the current process. Trying to take @msaboff's feedback
into account here.

### DIFF
--- a/index.html
+++ b/index.html
@@ -107,6 +107,8 @@
         <li>Complete spec text
         <li>Designated reviewers have signed off on the current spec text
         <li>The ECMAScript editor has signed off on the current spec text
+        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, in a PR
+        <li>If possible, there is an implementation in a polyfill or transpiler
       </ul>
     </td>
     <td>The solution is complete and no further work is possible without implementation experience, significant usage and external feedback.
@@ -121,9 +123,8 @@
     <td>
       <ul>
         <li>Above
-        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, and merged
-        <li>Two compatible implementations which pass the acceptance tests
-        <li>Significant in-the-field experience with shipping implementations, such as that provided by two independent VMs
+        <li>Test262 tests are merged
+        <li>Two complete browser implementations, shipped in a preview release, which pass the test262 tests
         <li>A pull request has been sent to <a href="https://github.com/tc39/ecma262">tc39/ecma262</a> with the integrated spec text
         <li>The ECMAScript editor has signed off on the pull request
       </ul>
@@ -172,7 +173,7 @@
 
 <h2>Test262 tests</h2>
 
-<p>During stage 3, <a href="https://github.com/tc39/test262">test262</a> tests should be authored and submitted via pull request. Once it has been appropriately reviewed, it should be merged to aid implementors in providing the feedback expected during this stage.
+<p>During Stage 2, <a href="https://github.com/tc39/test262">test262</a> tests should be authored and submitted via pull request. Once it has been appropriately reviewed and Stage 3 is reached, it should be merged to aid implementors in providing the feedback expected during this stage.
 
 <h2>Eliding the process</h2>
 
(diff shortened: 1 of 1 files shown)


mathiasbynens:
Thanks for tackling this, Dan. 👍 to merging this as it clarifies the _existing_ process. #13 warrants more discussion and can potentially be merged later.

domenic:
Strong -1. This does not clarify the existing process, but instead changes it substantially. This change does not have consensus.

index.html
```diff
@@ -181,3 +182,13 @@
...
+
+<h2>Browser implementations</h2>
+
+Browser implementations refer to web browsers such as Chromium, WebKit, Edge or Firefox. Implementations in web browsers often have additional constraints that other implementations don't have, such as a detailed understanding of web compatibility that comes from implementation so their experience is uniquely valuable. Specifically enfrancising browsers here, and differentiating them from transpilers, has the important benefit of making sure the committee doesn't add things to the specification that will never ship in the browser consensus.
```
ljharb, line 188: Chromium and WebKit seem like they're more in the same bucket as ChakraCore and SpiderMonkey; whereas Chrome and Safari would be in the same bucket as Edge or Firefox?

Also, what about non-transpiler non-browser native engines?

index.html
```diff
@@ -181,3 +182,13 @@
...
+
+For Stage 4, this process only requires complete implementations delivered in a preview release in a browser, and not shipping in stable releases. Here, a preview release refers to a mechanism by which users can opt into newer, unstable versions of a browser. In some, this may be triggered by turning on an "advanced JavaScript features" flag in browser settings; in others, this may be done by using a particular "tech preview", "nightly" or "insider edition" version.
+
+<h2>Polyfill/transpiler implementations</h2>
+
+Implementations in polyfills or transpilers such as Babel, TypeScript, core.js and es6shim help with working out kinks in the specification and getting initial user feedback. These implementations are specifically invoked by users, rather than being present by default in the execution environment. They are somewhat complementary to native implementations, as they are lighter weight to develop and change, so they are requested at an earlier stage in the process.
```
ljharb, line 194: `es6-shim`, and Oxford comma before `core.js`

index.html
```diff
@@ -107,6 +107,8 @@
         <li>Complete spec text
         <li>Designated reviewers have signed off on the current spec text
         <li>The ECMAScript editor has signed off on the current spec text
+        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, in a PR
```
ljharb, line 110: Just calling out that this is a specific change to the existing process

index.html
```diff
@@ -107,6 +107,8 @@
...
+        <li>If possible, there is an implementation in a polyfill or transpiler
```
ljharb, line 111: This is fine; stage 2 already conceptually means "this is when non-native implementations should ship"

index.html
```diff
@@ -121,9 +123,8 @@
     <td>
       <ul>
         <li>Above
-        <li><a href="https://github.com/tc39/test262">Test262</a> acceptance tests have been written for mainline usage scenarios, and merged
-        <li>Two compatible implementations which pass the acceptance tests
-        <li>Significant in-the-field experience with shipping implementations, such as that provided by two independent VMs
+        <li>Test262 tests are merged
+        <li>Two complete browser implementations, shipped in a preview release, which pass the test262 tests
```
ljharb, line 127: Enfranchising browsers here does have the effect of *dis*enfranchising non-browser implentations. I suspect that this will be controversial.

littledan:
> Strong -1. This does not clarify the existing process, but instead changes it substantially. This change does not have consensus.

I apologize for the poorly chosen name for this PR. Of course I'm making a proposal here, not saying that this is the current process.

msaboff:
> - Require test262 tests out for review for Stage 3
> - Require a polyfill/transpiler implementation for Stage 3
> - Require native browser preview release, rather than "shipping experience"

These seem to be more substantive changes than what I think is needed and does alter the existing process.

I agree with @ljharb that we shouldn't single out browsers over other implementations.  I wasn't present when this process was defined, but it appears that the desire here is for at least two independent implementation can implement the feature and agree in the fine details of the feature.  Those implementations and some of their users have some real world experience and recommend that the feature should be included in the standard.

The only item of the existing Stage 4 Entrance Criteria I think we need to discuss is **Significant in-the-field experience with shipping implementations, such as that provided by two independent VMs**.  I think it boils down to what kind of **shipping implementation** is sufficient for **significant in-the-field experience**.

index.html
littledan: Yep, sorry that the commit message was unclear in saying that this is a change.

littledan:
@msaboff What kind of shipping implementation do you think should be sufficient for significant in-the-field experience?

index.html
littledan: For context, I wanted to specifically give a role to transpilers/polyfills, if we're also specifically calling out native implementations for Stage 4; there's no intention to ignore them here. That's what's motivating the proposal for a change here.

msaboff:
@littledan My position is that a preview version of a browser is sufficient.  I understand that @domenic and others want the bar to be set higher at a released browser intended for wide consumer use.

This is the only discussion I think we need to have about moving to stage 4.

littledan:
OK, I'll follow up with a more targeted PR which just focuses on that aspect (to encode what @msaboff has described, which seems to be widespread in the committee).

PR #20, opened 2018-05-23
fixes #19: support a plurality of editors

### LINKED ISSUE #19: clarify stage 3 and 4 entrance criteria around editor approval
Now that we have an editor group instead of a single editor, I am unsure whether every member of that group must approve spec text (in the case of stage 3 entrance) or pull requests (in the case of stage 4 entrance). Are we to assume that any member of the editor group always speaks for the group?

Also, in the case of editor changes, does the approval of someone who was the editor at the time of the approval count as editor approval? What about approval of someone who was not the editor at the time of approval but is now the editor or a member of the editor group? I ask this not simply to explore the theoretical dilemmas but because I have already run into this with one of my proposals. And given we now have an editor group, I expect editor role changes to be more common than they were in the past.

Mentions of current and former editors for their input: @allenwb @bterlson @ljharb @bmeck

michaelficarra:
Fixes #19.

### DIFF
--- a/index.html
+++ b/index.html
@@ -106,7 +106,7 @@
         <li>Above
         <li>Complete spec text
         <li>Designated reviewers have signed off on the current spec text
-        <li>The ECMAScript editor has signed off on the current spec text
+        <li>All ECMAScript editors have signed off on the current spec text
       </ul>
     </td>
     <td>The solution is complete and no further work is possible without implementation experience, significant usage and external feedback.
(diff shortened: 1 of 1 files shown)


ljharb (approved):

littledan (approved):
Seems good to me to codify the editor groups' stated practice of internal consensus.

bmeck (approved):

PR #23, opened 2018-09-28
Document repository transfers

gibson042:
Transfer of repositories to TC39 should be documented as part of the proposal process.

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,9 +28,15 @@
 
 <p>The Ecma <a href="https://www.ecma-international.org/memento/tc39.htm">TC39</a> committee is responsible for evolving the ECMAScript programming language and authoring the specification. The committee operates by consensus and has discretion to alter the specification as it sees fit. However, the general process for making changes to the specification is as follows.
 
-<h2>Development</h2>
+<h2>Input into the process</h2>
+
+<p>Ideas for evolving the ECMAScript language are accepted in any form. Any discussion, idea, or proposal for a change or addition which has not been submitted as a formal proposal is considered to be a “strawman” (stage 0) and has no acceptance requirements. Such submissions must either come from TC39 delegates or from non-delegates who have <a href="https://tc39.github.io/agreements/contributor">registered</a> via Ecma International.
+
+<h2>Stages</h2>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -28,9 +28,15 @@ <h1>The TC39 Process</h1>
...
+
+<h2>Stages</h2>
+
+<p>There are five stages of maturity in the development of an idea into a fully specified feature, complete with acceptance tests and multiple implementations. The TC39 committee must approve acceptance for each stage, and uses the below guidelines to control advancement.
 
-<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are five stages: a strawman stage, and 4 “maturity” stages. The TC39 committee must approve acceptance for each stage.
+<p>Proposals at stage 1 and higher should be under the control of the TC39 committee. Upon proposal acceptance, any externally-controlled repositories should be transferred to the <a href="https://github.com/tc39-transfer">TC39 Transfers GitHub organization</a>, and from there to the <a href="https://github.com/tc39">TC39 GitHub organization</a> by a TC39 committee member.
```
gibson042: There seems to be a gap in how a repository gets to tc39-transfer in the first place, since arbitrary GitHub users do not have create access and knowledge about how to request it is basically distributed by informal channels rather than official documentation.

ljharb: These instructions already live here: https://github.com/tc39/proposals#onboarding-existing-proposals

Let’s keep that as the source of truth.

ljharb (changes requested):

index.html
mathiasbynens: Hmm… It seems weird for this to not be mentioned in the process document, despite it clearly being part of the process. We can certainly cross-link, but documenting this requirement here is a net gain IMHO.

ljharb: I’m fine with that; I’d expect it to link there, as you said.

gibson042: Done. Thanks for the link.

ljharb (approved):

ljharb:
Tweaked to use "delegate" instead of "member", and "owned" instead of "controlled".

Since this has been up approaching 2 years without objections, I'll land this, and we can revert or iterate in a followup PR.

PR #28, opened 2020-04-27
Remove duplication of "Input into the process" section

chicoxyzzy:
...to align with previous change https://github.com/tc39/process-document/commit/112792e2eb34a7b143a05fa7b81bb4918e5a7152


littledan (approved):

ryzokuken (approved):
Thanks for this change.

chicoxyzzy:
I just noticed that there are two "Input into the process" sections now. I think that we need to leave the latter, but maybe apply a change "member" -> "delegate" which was made in former.

ljharb (approved):

PR #31, opened 2021-07-14
Remove stage name column

gesa:
As decided by consensus on day 2 of the 84th meeting of the committee 14 July 2021 @ 13:30 JST


ljharb (approved):

leobalter:
Do we mention these respective names or anything like that on the --how we work or other-- TC39 projects?

Edit: we don't have strawman or strawperson on the how-we-work repo.

PR #36, opened 2023-10-13
add back stage names

michaelficarra:
As discussed at the most recent TC39 meeting. We previously removed stage names in https://github.com/tc39/process-document/pull/31 because we didn't like the names, but it seems like there was appetite for adding names back as long as they were more appropriate. Names taken from the examples in my presentation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -50,7 +50,7 @@
   </thead>
   <tr>
     <td>0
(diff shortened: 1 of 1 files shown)


michaelficarra:
This has been superseded by #38.

PR #37, opened 2023-11-16
add a new testing stage

michaelficarra:
To be used as a basis for the ongoing discussion of a new stage at the upcoming TC39 meeting.

### DIFF
--- a/index.html
+++ b/index.html
@@ -97,20 +97,34 @@
     <td>Experimental
   </tr>
   <tr>
-    <td>3
-    <td>Indicate that further refinement will require feedback from implementations and users
+    <td>TBD
+    <td>Gain initial experience with the feature through testing and prototyping
     <td>
       <ul>
         <li>Above
         <li>Complete spec text
         <li>Designated reviewers have signed off on the current spec text
-        <li>All ECMAScript editors have signed off on the current spec text
+        <li>ECMAScript editors have signed off on the current spec text
+      </ul>
+    </td>
+    <td>The solution is complete and no further work is possible without feedback from tests, implementations, or usage. No changes to the proposal will be requested aside from those elicited through testing or implementation experience.
+    <td>Complete: all semantics, syntax and API are completely described
+    <td>Limited: only those resulting from new information obtained through testing or usage experience
+    <td>Spec compliant prototypes
+  </tr>
+  <tr>
+    <td>3
+    <td>Indicate that further refinement will require feedback from implementations or users
+    <td>
+      <ul>
+        <li>Above
+        <li>The feature has sufficient testing and appropriate pre-implementation experience
       </ul>
     </td>
     <td>The solution is complete and no further work is possible without implementation experience, significant usage and external feedback.
     <td>Complete: all semantics, syntax and API are completely described
     <td>Limited: only those deemed critical based on implementation experience
-    <td>Spec compliant
+    <td>Spec compliant production
   </tr>
   <tr>
     <td>4
--- a/index.html
+++ b/index.html
@@ -97,7 +97,7 @@
     <td>Experimental
   </tr>
   <tr>
-    <td>TBD
+    <td>2.7
     <td>Gain initial experience with the feature through testing and prototyping
     <td>
       <ul>


index.html
```diff
@@ -91,26 +91,40 @@ <h2>Stages</h2>
         <li>Initial spec text
       </ul>
     </td>
-    <td>The committee expects the feature to be developed and eventually included in the standard
+    <td>The committee expects the feature to be developed and eventually included in the standard. No changes to the proposal will be requested aside from those elicited through testing or implementation experience.
```
ljharb: this applies to the new stage, i think, but definitely does not apply to stage 2, in which entire proposals have changed, often.

michaelficarra: Oops, I think I edited the wrong line. Will fix.

index.html
```diff
@@ -97,20 +97,34 @@ <h2>Stages</h2>
...
+      <ul>
+        <li>Above
+        <li>The feature has sufficient testing and appropriate pre-implementation experience
       </ul>
     </td>
     <td>The solution is complete and no further work is possible without implementation experience, significant usage and external feedback.
     <td>Complete: all semantics, syntax and API are completely described
     <td>Limited: only those deemed critical based on implementation experience
-    <td>Spec compliant
+    <td>Spec compliant production
```
ljharb, line 127:
```suggestion
    <td>Spec compliant; safe to use in production
```
maybe?

ljharb (approved):

index.html
michaelficarra: Since we got consensus on specifically this wording, I'm going to leave it alone. I'll be submitting a follow-up PR soon with names and a refactoring of descriptions that maintains the spirit here, which will include addressing this wording.

michaelficarra:
This should be good to merge as-is.

index.html
```diff
@@ -97,20 +97,34 @@ <h2>Stages</h2>
...
     <td>
       <ul>
         <li>Above
         <li>Complete spec text
         <li>Designated reviewers have signed off on the current spec text
-        <li>All ECMAScript editors have signed off on the current spec text
+        <li>ECMAScript editors have signed off on the current spec text
+      </ul>
+    </td>
+    <td>The solution is complete and no further work is possible without feedback from tests, implementations, or usage. No changes to the proposal will be requested aside from those elicited through testing or implementation experience.
```
bakkot, line 110:
```suggestion
    <td>The solution is complete and no further work is possible without feedback from tests, implementations, or usage. No changes to the proposal will be requested aside from those elicited through testing, implementation, or usage experience.
```
surely? given that "usage" is mentioned in the previous sentence and multiple times below.

index.html
```diff
@@ -97,20 +97,34 @@ <h2>Stages</h2>
...
+    <td>Complete: all semantics, syntax and API are completely described
+    <td>Limited: only those resulting from new information obtained through testing or usage experience
```
ctcpip, line 112: per bakkot's comment above, this needs updating to match that suggestion if accepted

index.html
ctcpip: nit: put implementation before testing

index.html
```diff
@@ -97,20 +97,34 @@ <h2>Stages</h2>
...
+    <td>Spec compliant prototypes
```
ctcpip, line 113:
```suggestion
    <td>Spec-compliant prototypes
```

index.html
ctcpip, line 127:
```suggestion
    <td>Spec-compliant production
```

ctcpip (approved):
LGTM, though the minor suggestions would be nice and perhaps consider amending the commit to not include the eyeroll emoji  🤷‍♀️

leobalter:
Ideas for a subsequent change (another PR), with the repetition of "_testing, implementation, or usage experience_" we might set these as a definition for "feature adoption" (name is up for bikeshed), so we only repeat the shorter verson.

E.g.: 

```diff
-The solution is complete and no further work is possible without feedback from tests, implementations, or usage. No changes to the proposal will be requested aside from those elicited through testing, implementation, or usage experience.
+The solution is complete and no further work is possible without feedback from tests, implementations, or usage. No changes to the proposal will be requested aside from those elicited through feature adoption.

+Feature adoption for proposals comprehends the experience observed by testing results, implementation, and/or usage.
```

This change must not block the progress/acceptance of this PR as it only serves as cosmetic/editorial.

michaelficarra:
These suggestions are great, but for process document changes, I don't feel comfortable merging anything other than the exact wording that was agreed upon. Unlike with how the committee delegates authority to the editor group to execute what they understand to be the committee's will, changes to this document are approved explicitly. Since I'm making a follow-up anyway, I'll incorporate these suggestions into that. But next time, *please* review the PR *before* it is approved.

PR #38, opened 2023-12-05
refactor the proposal stages table for better communication

michaelficarra:
As we discussed in plenary, this is an incorporation of the presented "purpose", "status", and "external communication" terms into the stage table of the process document. Additionally, the table is simplified by reducing to 4 columns from 7 without losing any information. The "status" column is targeted at the community and other external consumers, and the entrance criteria and purpose columns are targeted at champions and other TC39 representatives. Help text for the "purpose" column clarifies that "These are the actions that will be undertaken by the committee and the proposal authors while the proposal is in this stage".

While it is mostly a rewording, there are some "normative" changes made that I think more accurately capture how we use our process in practice:

* Moved "Proposal document describes all high-level APIs and syntax" and "Illustrative examples of usage" from stage 1 entrance criteria to stage 2 entrance criteria.
* Replace references to "ECMAScript editors" and "ecma262" with "the relevant editor group" and "ecma262 or ecma402" so that the document applies correctly to Intl proposals
* Removed test262 entrance criteria from stage 4 since it's now already required for stage 3.

Since it's been mostly rewritten, I'd recommend reviewing the rendering instead of the diff.

Here's a rendering of the table as of submitting this PR:

image

Ping @waldemarhorwat since he asked to be notified when I opened this PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -98,6 +98,7 @@
         <li>Work out minor details such as ordering of observable effects, handling of invalid inputs, API names, etc.
         <li>Receive and address spec text reviews from the assigned reviewers and the appropriate editor group
         <li>Produce experimental implementations such as loosely-correct (not for production use) polyfills to aid in validating the design and exploring the details
+        <li>Investigate integration with relevant host APIs, if necessary
       </ul>
     </td>
   </tr>
--- a/index.html
+++ b/index.html
@@ -30,7 +30,7 @@
 
 <h2>Stages</h2>
 
-<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are six stages: a strawperson stage, and 5 “maturity” stages. The TC39 committee must approve acceptance for each stage.
+<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are six stages: a strawperson stage, and five “maturity” stages. The TC39 committee must approve acceptance for each stage.
 
 <p>Proposals at stage 1 and beyond should be owned by the TC39 committee. Upon proposal acceptance, any externally-owned repositories should be transferred by following the <a href="https://github.com/tc39/proposals#onboarding-existing-proposals">onboarding instructions</a>.
 
--- a/index.html
+++ b/index.html
@@ -30,7 +30,7 @@
 
 <h2>Stages</h2>
 
-<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are five stages: a strawperson stage, and 4 “maturity” stages. The TC39 committee must approve acceptance for each stage.
+<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are six stages: a strawperson stage, and 5 “maturity” stages. The TC39 committee must approve acceptance for each stage.
 
 <p>Proposals at stage 1 and beyond should be owned by the TC39 committee. Upon proposal acceptance, any externally-owned repositories should be transferred by following the <a href="https://github.com/tc39/proposals#onboarding-existing-proposals">onboarding instructions</a>.
 
--- a/index.html
+++ b/index.html
@@ -48,7 +48,7 @@
 
   <tr>
     <td>0
-    <td>This is a new proposal. It is not currently being considered by committee.
+    <td>This is a new proposal. It is not currently being considered by the committee.
     <td>None. New proposals are assigned this stage by their authors outside of the usual advancement process.
     <td>
       Ideation and exploration. Define a problem space in which the committee and the champions can focus their efforts.
@@ -97,7 +97,7 @@
       <ul>
         <li>Work out minor details such as ordering of observable effects, handling of invalid inputs, API names, etc.
         <li>Receive and address spec text reviews from the assigned reviewers and the appropriate editor group
-        <li>Produce experimental implementations such as loosely-correct polyfills to aid in validating the design and exploring the details
+        <li>Produce experimental implementations such as loosely-correct (not for production use) polyfills to aid in validating the design and exploring the details
       </ul>
     </td>
   </tr>
--- a/index.html
+++ b/index.html
@@ -84,7 +84,7 @@
 
   <tr>
     <td>2
-    <td>The committee has chosen a preferred solution or solution space, but the design is a draft and may still change significantly. The committee expects the feature to be developed and eventually included in the standard, but due to reasons that may not be apparent at this stage, the feature may never advance.
+    <td>The committee has chosen a preferred solution or solution space, but the design is a draft and may still change significantly. The committee expects the feature to be developed and eventually included in the standard, but due to reasons that may not be apparent at this stage, the feature may never be included in the standard.
     <td>
       <ul>
         <li>Proposal document describes all high-level APIs and syntax
--- a/index.html
+++ b/index.html
@@ -36,117 +36,115 @@
 
 <table>
   <caption>ECMAScript Proposal Stages</caption>
+
   <thead>
     <tr>
       <th>Stage
-      <th>Purpose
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -36,118 +36,116 @@ <h2>Stages</h2>
...
-    <td>Allow input into the specification
-    <td>None
-    <td>N/A
-    <td>N/A
-    <td>N/A
-    <td>N/A
-  </tr>
-  <tr>
-    <td>1
+    <td>This is a new proposal. It is not currently being considered by committee.
```
ljharb:
```suggestion
    <td>This is a new proposal. It is not currently being considered by the committee.
```

index.html
```diff
@@ -36,118 +36,116 @@ <h2>Stages</h2>
...
+        <li>Illustrative examples of usage
+        <li>Initial spec text including all major semantics, syntax, and APIs. Placeholders, TODOs, and editorial issues are acceptable
+      </ul>
+    </td>
+    <td>
+      Refining the solution.
+      <ul>
+        <li>Work out minor details such as ordering of observable effects, handling of invalid inputs, API names, etc.
+        <li>Receive and address spec text reviews from the assigned reviewers and the appropriate editor group
+        <li>Produce experimental implementations such as loosely-correct polyfills to aid in validating the design and exploring the details
```
ljharb:
```suggestion
        <li>Produce experimental implementations such as loosely-correct non-production-usage polyfills to aid in validating the design and exploring the details
```

ljharb (changes requested):
Overall this looks great and seems to preserve all the criteria and text we've relied on, nice work.

index.html
michaelficarra: Accepted.

index.html
michaelficarra: Accepted as "loosely-correct (not for production use) polyfills".

ljharb: That captures what I wanted, thanks :-)

ljharb (approved):
LGTM. If there's any of the wording in #33 that seems relevant here, that would increase the chances of obviating that PR entirely :-)

michaelficarra:
@ljharb I thought about it, but since I couldn't entirely obviate it (I would love to see that W3C document referenced), I decided not to go that route.

index.html
```diff
@@ -30,124 +30,122 @@ <h1>The TC39 Process</h1>
 
 <h2>Stages</h2>
 
-<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are five stages: a strawperson stage, and 4 “maturity” stages. The TC39 committee must approve acceptance for each stage.
+<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are six stages: a strawperson stage, and 5 “maturity” stages. The TC39 committee must approve acceptance for each stage.
```
ctcpip:
```suggestion
<p>Changes to the language are developed by way of a process which provides guidelines for evolving an addition from an idea to a fully specified feature, complete with acceptance tests and multiple implementations. There are six stages: a strawperson stage, and five “maturity” stages. The TC39 committee must approve acceptance for each stage.
```

index.html
```diff
@@ -30,124 +30,122 @@ <h1>The TC39 Process</h1>
...
     </td>
-    <td>None
-    <td>Major
-    <td>Polyfills / demos
   </tr>
+
   <tr>
     <td>2
-    <td>Precisely describe the syntax and semantics using formal spec language
+    <td>The committee has chosen a preferred solution or solution space, but the design is a draft and may still change significantly. The committee expects the feature to be developed and eventually included in the standard, but due to reasons that may not be apparent at this stage, the feature may never advance.
```
ctcpip: > but due to reasons that may not be apparent at this stage, the feature may never advance.

wording sounds a little strange to me, but I don't immediately have a suggestion to improve

index.html
michaelficarra: Accepted.

index.html
michaelficarra: Well, this isn't really actionable for me. The wording doesn't sound strange to me, so until you have a suggested alternative, marking as resolved.

ctcpip: > reasons that may not be apparent at this stage

if the proposal will not advance to the next stage, that suggests that the reasons _will_ be apparent at this stage

ctcpip (approved):

index.html
ctcpip: so thinking about this a bit more, is what we're really saying here: due to reasons that may not be apparent at this stage (stage 2), the feature may never advance (to stage 3), and it first needs to go to stage 2.7 for those reasons to become apparent.  (edit: or those potentially blocking reasons did not manifest at stage 2.7, so it may advance to stage 3)

I realize this is still not a concrete suggestion for a change, but does that track?

michaelficarra: That is one possibility. The original wording that I was trying to preserve was "The committee expects the feature to be developed and eventually included in the standard". Since this column is intended to target the community, I wanted to moderate the wording here as to not convey a commitment (through the use of "expects") to proposals at stage 2. Anything short of making it to stage 4 could be seen as a failure of this commitment. So I modified it to what you see here.

ctcpip:
```suggestion
    <td>The committee has chosen a preferred solution or solution space, but the design is a draft and may still change significantly. The committee expects the feature to be developed and eventually included in the standard, but due to reasons that may not be apparent at this stage, the feature may never be included in the standard.
```

?

michaelficarra: Sure, done.

index.html
```diff
@@ -36,118 +36,116 @@ <h2>Stages</h2>
...
+        <li>Illustrative examples of usage
+        <li>Initial spec text including all major semantics, syntax, and APIs. Placeholders, TODOs, and editorial issues are acceptable
+      </ul>
+    </td>
+    <td>
+      Refining the solution.
+      <ul>
+        <li>Work out minor details such as ordering of observable effects, handling of invalid inputs, API names, etc.
+        <li>Receive and address spec text reviews from the assigned reviewers and the appropriate editor group
+        <li>Produce experimental implementations such as loosely-correct (not for production use) polyfills to aid in validating the design and exploring the details
```
bakkot, line 100: Maybe add
```
        <li>Investigate integration with relevant host APIs, if necessary
```
or something like that. See the recent ShadowRealms demotion.

bakkot (approved):
LGTM though ideally I'd like to see something about integration with hosts, as mentioned above

index.html
michaelficarra: Sure! Done.

ctcpip (approved):

michaelficarra:
Since the initial wave of feedback seems to have concluded, the PR has stabilised, and we've received some approvals, I've generated an updated rendering to make it easier for any further reviewers to review the current revision:

image

ljharb:
All of them have kind of a "headline" under the Purpose column except 3 and 4, should they have them too? Perhaps "Implementing and Shipping" and "Specifying and Merging" or something?

michaelficarra:
I figured those ones are kind of *just* the "headline". Stage 0 purpose also has a bit of a longer "headline" FWIW.

littledan (approved):
This is an excellent change. +1 to merging as is.

Note that I wrote related documentation at https://github.com/tc39/how-we-work/blob/main/champion.md#moving-through-the-stages-in-committee. If it becomes awkward to put too much text in this table, we can consider long-form like that. But it is certainly valuable that you ensured that this stuff is in the process document and has committee consensus, which I did not do.

Some further improvements to consider (whether in this patch or later):
- The help text for "Purpose" is great, but I still worry about people who don't see it. Maybe the headline could be something like "Work during stage"?
- It's great that this is framed in bullet points largely, but consider using more things to break this up visually to avoid looking like a blob of text, e.g., bold, more line breaks, spacing, etc.
- The change removing the quotes around "champion" reminds me that we could cross-reference this with our glossary! just a thought.
- As @ljharb suggests, some two-word summaries of what the stages mean would be valuable. I think the stage titles that you proposed would go perfectly here--including them would be consistent with the feedback from committee that numbers should remain primary, but the descriptions are still useful.

michaelficarra:
This has reached consensus at today's plenary.

PR #42, opened 2024-03-07
even out the column widths in the stages table

michaelficarra:
This makes for better reading on a wide display.

before:

image

after:

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,6 +16,14 @@
     border-collapse: collapse;
   }
 
+  #stages {
+    table-layout: fixed;
+    width: 100%;
+  }
+  #stages th:first-child {
+    width: 55px;
+  }
+
   ul {
     padding: 0 1.5em;
     margin: 0;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -16,6 +16,14 @@
     border-collapse: collapse;
   }
 
+  #stages {
+    table-layout: fixed;
+    width: 100%;
+  }
+  #stages th:first-child {
```
ljharb, line 23: there's only one table, so presumably this doesn't need the ID at all, and could be combined with the existing `table { }` declaration?

michaelficarra: I guess but that kinda leaves a hazard if we ever add another table. I don't see a reason not to just say the thing we mean.

ljharb: then should the other one be changed to reflect the ID?

michaelficarra: It is plausible that the author intended that to apply to all tables. I think it's a common "default override".

ljharb (approved):

PR #50, opened 2025-01-22
♿️ a11y - text spacing and alignment fixes, add dark mode

ctcpip:
## issues addressed

- WCAG Success Criterion 1.4.12 Text Spacing
  - note, two of the rules for compliance are currently commented out.  CSS letter spacing and word spacing makes the text look quite weird
- ~converted the stage table to regular layout.~ "[Tables must not be used as layout aids.](https://html.spec.whatwg.org/multipage/tables.html#the-table-element:~:text=Tables%20must%20not%20be%20used%20as%20layout%20aids.)"
  - also they are not great for a11y and small displays
  - this change has been backed out.  (see comments below)
- added dark mode for those who celebrate
- there are no changes to text content

### befores and afters

Screenshot 2025-01-22 at 4 01 57 PM

### dark mode

Screenshot 2025-01-22 at 4 02 23 PM

### validations

Screenshot 2025-01-22 at 3 13 02 PM

Screenshot 2025-01-22 at 3 30 08 PM

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,12 +5,23 @@
 <style>
   html {
     font-family: sans-serif;
+    font-size: 16px;
+
+    /* WCAG 2.1 - Line height (line spacing) to at least 1.5 times the font size */
+    line-height: 1.6;
+
+    /* WCAG 2.1 - Letter spacing (tracking) to at least 0.12 times the font size */
+    /* letter-spacing: 0.12em; */
+
+    /* WCAG 2.1 - Word spacing to at least 0.16 times the font size */
+    /* word-spacing: 0.16em; */
   }
 
   th, td {
     vertical-align: top;
     border: 1px solid black;
     padding: 0.4em;
+    text-align: left;
   }
 
   table {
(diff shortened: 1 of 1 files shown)


ljharb:
We're not using tables for layout here, we're using it to display tabular data - I think the use of tables is thus correct and should remain.

ctcpip:
- that's not tabular data
- it's poor for a11y
- it's poor on mobile

bakkot:
(I'm confused by the claim that this is not tabular data - this is data with two dimensions, namely, stage numbers and properties of stages. If this isn't tabular data, I don't know what is.)

I don't know enough about accessibility to comment on whether or not the new layout is better on that metric, but it's definitely harder for me personally to use, so I'd want to be sure this was actually an improvement for other people before landing it.

For mobile, any chance we can do one of those fancy tables that collapses when the page width is too small, instead of doing away with the table entirely?

ljharb:
It is definitely tabular data, and I'm not sure why it's poor for a11y - screen readers have long been able to navigate tables well, afaik.

As far as poor on mobile, the existing page seems to work fine on my iPhone? The text is initially small, but that's what pinch zoom is for.

ctcpip:
I suppose I interpret tabular data as typically structured data, rather than unstructured text.  e.g. something you could plot on a graph.

When tables are used for text content, screen readers may misrepresent the information, making it confusing or inaccessible.

In any case since it sounds like folks really like the table, I'll see about what we can do to help with usability on small displays.

ctcpip:
I'll need more time to experiment with the table display.  For now, I've backed out the table change, so this is now only the text spacing and bonus dark mode.

bakkot (approved):

sffc:
On the question of whether or not to use a table, one of my teammates in a11y said: "for me it usually comes down to a judgment call primarily based on the amount of data and the nature of the logical relationship of aspects of the data ... largely situational and more of a UX choice than something that has a strict guideline."

michaelficarra (approved):

